PR #32, opened 2015-02-16
changed ny to n1 and nx to n0 so that the labels match the text, it's qu...

psteinb:
...ite confusing otherwise
It's a great library!


stevengj:
Thanks!

PR #40, opened 2015-05-02
add new switch '--with-only-r2hc' to use r2r API only, this switch will ...

liangyaozhan:
...reducing code size.


stevengj:
See the wisdom2conf program for reducing code size.

PR #53, opened 2016-01-20
Fixed unique token generation in genfft

artemkin:
OCaml became smart enough to break this trick. It inlines `make_aux` starting from 4.02.0, and all generated tokens are the same.

```ocaml
(* this depends on the compiler not being too smart *)
let make () =
  let make_aux x = Unique x in
  make_aux ()

(* note that the obvious definition
      let make () = Unique ()
   fails *)
```

It seems this led to generation of wrong codelets (at least genfft built with OCaml 4.01.0 and 4.02.3 generated different codelets).

I've implemented "legal" unique token generation.

### DIFF
--- a/genfft/unique.ml
+++ b/genfft/unique.ml
@@ -21,18 +21,14 @@
 
 (* repository of unique tokens *)
 
-type unique = Unique of unit
+type unique = int
 
-(* this depends on the compiler not being too smart *)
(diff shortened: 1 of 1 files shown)


matteo-frigo:
Applied, thanks.  I have made a couple of stylistic changes.

PR #58, opened 2016-03-17
Update configure.ac

ruiaraujo:
While cross compiling for Android with NEON enabled it complained about the missing -mfloat option.

I am not really sure if this is fully correct but it fix my build.


rdolbeau:
NEON can be used on aarch64 and on hard-float arm ABIs as well, so the soft-float ABI shouldn't be forced in configure.

PR #69, opened 2016-08-31
Build und install cmake module

junghans:


junghans:
Also see, https://gitlab.kitware.com/cmake/cmake/merge_requests/77#note_159555

junghans:
@xantares: Did you added something like this to #103?

matteo-frigo:
I pulled #103

junghans:
@matteo-frigo: This PR isn't included in #103. (I just looked through it.)

matteo-frigo:
Ack.

I pulled your patch.  Does it deserve any documentation, or does everything happen automatically?  (I.e., does cmake know by default to look into /usr/local/share/cmake/Modules or whatever?)

junghans:
> I pulled your patch. Does it deserve any documentation, or does everything happen automatically? (I.e., does cmake know by default to look into /usr/local/share/cmake/Modules or whatever?)

Yes, everything is automatically.

I will send another PR to make FFTW's new cmake buildsystem generate the `FindFFTW3.cmake` file as well.

xantares:
That's not the advised way of writing cmake config files, look for config mode:
https://cmake.org/cmake/help/v3.0/command/find_package.html

Basically you should instead provide a FFTW3Config.cmake that's installed into (libdir)/cmake/fftw3
This file defines FFTW3_LIBRARIES, FFTW3_INCLUDE_DIRS, version etc

junghans:
@xantares: Yeah, but this PR was from before #103 and only meant to support project linking again fftw (build by autotools).

PR #80, opened 2017-01-15
[RFC|WIP] Proposal for full CMake support

EikeVerdenhalven:
Objective: Introducing a platform agnostic build system based on CMake

As FFTW seeks to build for many different architectures, I propose here to move the build system to CMake. With the implementation presented here it is already possible to configure the project and create all generated source files.
The generation requires an ocaml installation as well as a working version of ocamlbuild. Optionally the generated codelets can be formatted with clang-format. The respective tools can be located using cmake defines or the cmake gui tool.
Furthermore benchmarks, tests and wisdom tool are covered. The tests have been integrated to run with CTest.

However the support for building dynamic libraries (especially on Windows) remains open work. Therefore this PR is rather a request for comments to see if the maintainers are interested in integrating the authors work.


matteo-frigo:
Eike,

you will need to be much more persuasive about what problem you are trying to solve and about why we would want to switch.

The distribution mode for FFTW over the past twenty years has been that the user downloads a fftw tarball, types configure && make.  The user doesn't need ocaml, pthreads, indent, clang_format, nothing.  The user only needs /bin/sh, make, and a C compiler.   I routinely test fftw on all debian releases since potato, and on RHEL5 and later (although RHEL4 is probably still ok), and various *BSD.  Moreover, we have plenty of users on AIX, HPUX, whatever.  We have users that cross-compile from linux to vxworks, qnx, or whatever embedded system.  You can cross-compile from pretty much any system to any other system.  (That's how we ship windows binaries in the first place---we cross-compile from linux and we don't have to touch windows.)

So, if we switched to cmake, we would add a dependency on cmake for all existing users, which would piss everybody off.   Even if we were willing to piss everybody off, you would need to convince me that cmake runs on all platforms where make+sh runs, which I doubt since cmake is written in C++ and every time I try to compile a C++ program one more than one platform something goes wrong.  You would need to convince me that all the cross-compilation properties, and all the ugly things such as building shared libraries on AIX, are still supported.

The only way in which this can work is if we support both cmake and automake.  You would have to make the case that this is a good idea, e.g., because automake does not really integrate nicely with Windows.  That's a possible solution to a long-standing problem.  On the other hand, I don't see why free software developers ought to invest their time to support a proprietary operating system.

There is another issue.  As you have found out, users are not supposed to download code from github, in the sense that we have a "release" process where we  compile genfft via ocaml once on our machine, generate C code, and produce a tarball posted on fftw.org.  The user downloads the tarball and doesn't have to know how the sausage is made.  Of course we are not preventing anybody from running ocaml themselves, but we don't want to require it.  Consequently, there is no possible argument to switch the "make dist" process to cmake, since I am the only one running it, we have a process that is not broken, and we have to support Makefiles no matter what.  Thus, any cmake discussion would have to be about how to compile the generated code and not how to generate it.

hadim:
That support would be awesome for the conda-forge (https://conda-forge.github.io/) project and automatic fftw3 packaging : https://github.com/conda-forge/fftw-feedstock/issues/2

timblechmann:
@matteo-frigo as for the FUD about c++: cmake is very conservative about the c++ features which are allowed in its codebase to ensure portability.

otoh: is it currently possible to build static libraries for msvc with the current buildsystem? they cannot be cross-compiled and has they need to be built against the c runtime of the specific compiler. i've seen people who have recreated visual studio project files because of that.

matteo-frigo:
@timblechmann Yes, compilation on Windows is a long-standing problem.  But again, it is unreasonable to expect authors of free software to overhaul their entire build system so that their software works on Windows.  This has to be clear: if people insist on using Windows in this day and age, they get what they deserve, and it cannot possibly be my problem. We survived with this attitude for twenty years when refusal to touch Windows was considered borderline insane, and I am not going to change my attitude now that Unix has won.

That being said, we are willing to support cmake as an alternate build system, and maintain it going forward, and test cmake builds before we make new releases.  It's not a big deal as long as somebody else is willing to do the initial work.  There isn't any benefit in having cmake run ocaml, so I would not spend time on that.  It would be really really good if we could generate config.h.in automatically as it is today.  We can keep the problem tractable by saying that cmake covers configurations not already covered by the existing automake system---e.g., there is no need for cmake to support things like VSX on Power8.

cmake better take backward compatibility religiously.  The day I hear a complaint from a user that the CMakelists file doesn't work because the user has a newer cmake version than I have, the whole thing goes out of the window and I don't want to hear about cmake ever again.

timblechmann:
@matteo-frigo fwiw, this PR comes from a customer who is using fftw3 under a commercial license. as for cmake's policies on compatibility, https://cmake.org/Wiki/CMake/Policies covers most of it.

in the academic / scientific / server-centric world, free-software unix is strong. in the user-facing world the unix that is becoming very strong (darwin) can hardly be considered as free software, though. 

but lets leave the dogmatics in the church as too many people have been killed in the name of religion and focus on the technical issues and implications ;)

matteo-frigo:
Practically speaking, RHEL5 comes with cmake-2.6.4.  Is it practical to set up things so that all versions of cmake since 2.6.4 work out of the box?

stevengj:
The technical issue here is that `cmake` makes it much easier to build FFTW on Windows, at the price of requiring us to either maintain two completely independent build systems or to switch over to `cmake` completely (which requires a substantial effort to make sure that it reproduces all the features of our rather-complicated build system).

Those are both steep prices for us to pay, considering that Windows users can just use the pre-built FFTW binaries that we cross-compile (or they can install mingw).   Why do Windows users need to build FFTW themselves?

(I'm not so worried about the dependency on `cmake`, per se.  `cmake` is pretty common these days, and it wouldn't be terribly onerous to require people to install it in the nowadays-rare case where they need to compile FFTW themselves.  I don't particularly *like* `cmake` — I find its language to be mildly insane, but to be fair the same can be said for the sh+m4+make mixture that is `automake`.)

timblechmann:
> Those are both steep prices for us to pay, considering that Windows users can just use the pre-built FFTW binaries that we cross-compile (or they can install mingw). Why do Windows users need to build FFTW themselves?

there are no pre-built static libraries. there are certain use cases, where using shared libraries is not an option. without going too much into details, most of the issues arise from using a library X (e.g. fftw) from a shared library Y (client code), as windows does not have a notion of RPATH. even worse, libraries are resolved by name. there are certain cases, when you will run into undefined behaviour.
loading libraries at run-time via the win32 equivalent of dlopen/dlclose has it's own problems. the most robust way to solve all those problems is to use static libraries. static libraries produced by different compilers (and compiler versions) are incompatible.

for many users, cross-compiled dlls are fine. but there are use cases where dynamic linking causes lots of troubles. 


> (which requires a substantial effort to make sure that it reproduces all the features of our rather-complicated build system, and requires us to ditch a system that works for us personally and that we are familiar with)

10 years ago many more people where familiar with autotools than these days and those who where almost exclusively had a unix background. these days much more people are familiar with cmake. so more users may contribute than to an autotool based buildsystem.

stevengj:
Providing pre-built static libraries seems useful, but it seems like something we should be able to do without `cmake`.

stevengj:
It looks like it should be sufficient to just build the `.a` static-library with the MingGW cross-compiler script, since apparently [visual studio can link `.a` files just like `.lib` files](http://stackoverflow.com/questions/2096519/from-mingw-static-library-a-to-visual-studio-static-library-lib).

timblechmann:
@stevengj i'm more concerned about runtime library conflicts related to ucrt and user code defaulting to different calling conventions.

stevengj:
@timblechmann, as far as I know, the calling conventions are the same (they are standardized); the only variation is stdcall vs cdecl (for 32-bit systems), which determined by the declarations in the header file.  As is the CRT ABI, I think, so it should work with whatever CRT the executable links to.  (And in any case, this would affect the DLLs too, which people have been using for years with no trouble, no?)

timblechmann:
@stevengj msvc has 3 different calling conventions and the default can be configured as compiler flag: https://msdn.microsoft.com/en-us/library/46t77ak2.aspx. functions can be annotated to use a specific calling convention, but `fftw3.h` doesn't seem do to that.

as for the c run-time: dlls and static libraries work *completely* differently. when linking a dll, the run-time library is linked into the dll (at link time of the dll). but for static libraries (which is just a set of object files), the run-time library is linked when linking the static lib into the target (at link-time of the target).

stevengj:
@timblechmann, it would be straightforward to add a `__cdecl` decorator to fftw.h, and we should probably do that, although people seem to rarely run into this in practice because most people use the default.

With the CRT, I still don't see the problem.  Why wouldn't an FFTW static library work with any version of the CRT that the target links to?

We should try it, in any case.  (It could be that MingW does something screwy that causes problems, but it is worth a try, and much easier than switching to cmake.)

stevengj:
As an experiment, I cross-compiled 64-bit Windows static libraries with MinGW for master; you can download them at http://jdj.mit.edu/~stevenj/fftw-3.3.6-pl1-dll64.zip (the `.a` files are the static libraries).   If someone wants to try them out with MSVC, that would be great.

timblechmann:
@stevengj 

> Why wouldn't an FFTW static library work with any version of the CRT that the target links to?

short answer: different c runtimes may have different ABIs

long answer:
if you compile some code, which uses c standard headers, which are provided by the c standard library. e.g. one c library could define a certain function inline, but another may define in the run-time library. one you link it together, you may have an undefined symbol. this is exactly the reason, why you cannot link a static library compiled with msvc 2013 into a target compiled with msvc 2015 for example.

--

btw, does the current buildsystem ensure that structs are not padded? (compare the mingw compiler option -mms-bitfields)

stevengj:
The question is, does this actually happen for the portion of the C standard library that FFTW uses?   We basically just use math functions like `sin`, and I/O functions like `fopen` and `printf`.  Have these really changed in ABI?  Most of the comments I can find online about CRT conflicts are in reference to the C++ standard library, which is famously inconsistent in its ABI.

timblechmann:
@stevengj compare https://msdn.microsoft.com/en-us/library/bb531344.aspx, especially the section starting with "The printf and scanf family of functions are now defined inline."

EikeVerdenhalven:
Dear @matteo-frigo , dear @stevengj ,

I recognize, that maintaining a software project that puts emphasis on backwards compatibility for legacy platforms is not a simple or small task. The discussion about calling conventions in this PR just exposes a glimpse of the difficulties you can experience when dealing with closed-source pre-built binaries. Therefore the ability to seamlessly integrate the compilation of open-source projects is in my opinion incredibly valuable. As was already mentioned software that appeals to the common user (be it free or commercial) needs to target the predominating commercially available platforms. Vendors of these platforms provide out of the box working compilation toolchains and SDKs. CMake provides a confined component provided for these platforms and allows a convenient and reliable way to configure any given C/C++ software to use these toolchains natively (e.g. no cygwin, MinGW, et al. are required).

Putting the aforementioned aside I think at this day and age the concept of needing to release code through a manual generation process seems ... let's say: a bit 'outdated'. For open-source projects on GitHub various cloud-services like travis, appveyor et al. exist, that allow to run builds and deploy artifacts automatically. Be having the process outlined in the repository it makes the e.g. the source generation process more transparent. Why would you hide it from the user? 
If you're interested you might take a look at my fork and see travis CI in action, building and testing each commit.

As much as I am absolutely convinced that CMake is a more flexible tool than the autotools chain, I can see how such a (admittedly vast) change can cause work. The current implementation is imho mature for use on linux/darwin/win32 (the acceptance tests pass), but of course I can not guarantee each and every case for every compiler and platform out in the world. Nevertheless I can imagine setting up cloud-based CI tools could help a lot to ensure compatibility here.

A final note: I started implementing the CMake build system for the FFTW library because I think it is an outstanding piece of software (or even piece of art :) ). Trying to get the source code generation platform agnostic was one of the things that I though would be a nice exercise to explore what CMake can do and hone my skills.
I am happy to further assist in the integration (and will probably work on it, when the time allows).
In case no interest from the maintainers exists to integrate the PR (or maybe a later refined, rebased :) version), I'd nevertheless thank the maintainers for reviewing and I will keep the CMake build system on a separate fork. 

- Eike

bagong:
SuperCollider would be eternally grateful if fftw3 could be built with Visual Studio ;) We are hoping to switch in the nearer future. We depend on Qt, which seems to be fading out the MinGW build flavor...

I tried hard to reproduce the source generation on Windows, and was not able to install an Ocaml build environment that would do what was expected (if running bootstrap.sh is the only thing required in a properly working ocaml build environment).

bagong:
Libsndfile is getting there atm :smile:

bagong:
Hi @EikeVerdenhalven , I have been trying to make this useable in a VS2015 compile. Unfortunately I don't understand much of what I am doing, but woul you be willing to comment on these steps?:

- On OSX I used your unity-ci branch and followed the build-steps contained in the CI scripts
- I ended up with a generated, zipped source
- moved and extracted it to Windows
- built with cmake the standard way
- I added USE_SSE and USE_SSE2, but none of the other optimizations
- I wasn't sure about the threads thingy, should I build them? (maybe combined into a single library?). What does the decision depend on?

Thanks for any hint. While I am not sure whether what I did makes sense, it went amazingly well. The build with msvc is very clean...

Thanks for your efforts!!!

EikeVerdenhalven:
Hi @bagong ,
the steps you mentioned should definitely lead to a working lib on windows. I had the idea of providing the zipped source file as a release on github, but wasn't sure if it would be confusing to users or be an infringement of the original authors work.
You can enable as many optimization flags as you like, since FFTW will choose the best suitable at runtime.
In general the CMake flags are one-to-one replacements for the options provided by the original autotools system. In the same way the combined threads option creates a single library that contains the multi-threaded versions of the library.
Currently the CMake implementation lacks support for non-X86/intel platforms and MPI builds.

Cheers,

Eike

bagong:
Thanks, @EikeVerdenhalven !

My eyes keep contracting [RFC|WIP] to RIP. I hope they err!

Bringing this into the main repo would be very useful for many. It would also help avoiding the proliferation of subtle build variations established by hand-knitted build definitions like here:

https://github.com/Microsoft/vcpkg/tree/master/ports/fftw3

Cmake has come a long way and is lately developing into the quasi standard as meta-buildsystem for cross platform C/C++. Even MS has given in, as seen in vcpkg and VS2017. Shouldn't that be a nice role model? ;)

jakirkham:
Would add that there is much more value to CMake than just providing support for Windows (not to detract from that value at all). As CMake is able to generate various different types of build scripts (obviously `Makefile` included), it is also able to generate things like [Ninja]( https://ninja-build.org/ ) build scripts, which are incredibly handy for building large numbers of packages quickly. However writing Ninja scripts directly would be a painful and pointless exercise. So having CMake support for FFTW would make it easier to build FFTW faster in an increasingly build time limited ecosystems like [conda-forge]( http://conda-forge.github.io/ ).

MathiasMagnus:
Having read through the discussions, I wanted to add my 5 cents. Trying to minimize the religious part of my comments.

I am a physics PhD student doing numerical relativity and data analysis, mostly in latest C++ standards, but mostly being tied by GPGPU APIs (OpenCL, C++AMP, SYCL), but I also use OpenMP, MPI, and OpenGL(-CL interop) from time to time. All cross-platform technologies. My primary development platform is Windows with visual Studio, but because HPC clusters are all Linux, that is my target platform. CMake is the bread and butter of my daily routine. Visual Studio picked up [prime-time support](https://blogs.msdn.microsoft.com/vcblog/2016/10/05/cmake-support-in-visual-studio/) for CMake as a supported project definition language. (Support for Ninja back-end [just hit](https://blogs.msdn.microsoft.com/vcblog/2017/06/14/cmake-support-in-visual-studio-whats-new-in-2017-15-3-preview-2/) inside the IDE itself.)

> every time I try to compile a C++ program one more than one platform something goes wrong

Yes,  if people use MSVC/GNU extensions all over, it's bound to happen. ISO C++ is reasonably portable. ISO C has just as many pitfalls. Comparing C++ compilation on multiple platforms to compiling C with just GCC (and claiming that cross-compilation is equivalent to using native tools) is not without bias.

> So, if we switched to cmake, we would add a dependency on cmake for all existing users, which would piss everybody off.

Why? Do people build autoconf from source? No. They use their platforms package manager to install dependencies. Do you build ocaml from source? CMake is readily available on most platforms.

> if people insist on using Windows in this day and age, they get what they deserve, and it cannot possibly be my problem, and it cannot possibly be my problem. We survived with this attitude for twenty years when refusal to touch Windows was considered borderline insane, and I am not going to change my attitude now that Unix has won.

_(Rephrasing for the 5th time...)_ Let's just say in my book, [this](http://gs.statcounter.com/os-market-share#monthly-201704-201704-bar) is not "Unix has won". Or as it is often phrased: "Using a commercial operating system on gets what he/she deserves..." Windows licenses are provided by the Hungarian state (through a nation wide agreement spanning schools, universities and all academic entities), and as such it is just as much paid for through taxes as our healthcare system. Something being commercial is not the moral equivalent to child murder and free-software developers could really stop acting as if it were.

> Putting the aforementioned aside I think at this day and age the concept of needing to release code through a manual generation process seems ... let's say: a bit 'outdated'...

Couldn't agree more.

> The user downloads the tarball and doesn't have to know how the sausage is made.

I am generally not against people hiding complexity from me, but thinking that their way is the best is dangerous incentive. Thinking that a cross-compiled binary makes me happiest instead of using the [vecorizer and auto-parallelizer](https://channel9.msdn.com/Events/GoingNative/2013/Compilerpp-) of my choice is a long shot.


So, with all due respect, what kept the project alive for the past 20 years may keep it going for another 10-20 years, but not being open about contributions like this will drive developers to contribute to other projects, may that be clFFT or any other competing tech that is more native-tool friendly. I am currently trying to build an old(er) analysis framework written in pure C to validate a newer, GPU accelerated tool. Compiling natively using the full Visual C toolchain is great help for debugging.

matteo-frigo:
Just a couple of observations to set the record straight, since Mathias' comment is not 100% factually correct.

First, FFTW is 100% C89 and it compiles with every C compiler under the sun.  It's not "just gcc".   We consider it a bug if FFTW does not compile with a C89 compiler.  (The rationale for C89 vs. C99 is not because we are reactionary: for a long time MSVC did not support C99, and I am not sure it does even today.  We would use C99 if we could.)  Of course, one can enable machine-specific extensions (sse2, avx, etc.) that require compiler support, but one can run FFTW without those.

Second, there is a cognitive dissonance between the github way of doing things and the historical FFTW way of doing things, as I now explain.  This dissonance is the root of many misunderstandings.

When I say "user", I mean somebody who downloads source code and compiles it.  (Specifically, I do not mean somebody who obtains a binary from debian or fedora---that's debian's problem.)  The intended distribution model of FFTW is that the "user" downloads fftw-3.x.tar.gz from fftw.org, types configure && make install.  fftw-3.x.tar.gz is a source tarball, not a binary.  In order to compile the source tarball, the user does *not* need autoconf, ocaml, nothing.  FFTW depends on a Unix Bourne-shell (/bin/sh), an old-school "make" (not GNU make), and a C compiler, and nothing else.  The shell doesn't have to be /bin/bash; for example, the Korn shell in OpenBSD works fine.  The Solaris /bin/sh worked fine last time I tried.

The repository on github contains a small fraction of the "official" distribution.  In order to generate the "official" fftw-3.x.tar.gz given a github repository, one needs ocaml, autoconf, automake, libtool, and other stuff (e.g., GNU make).  The "outdated" process of releasing a source tarball ensures that the "user" does not have to obtain compatible versions of these tools.  For example, autoconf keeps changing and breaking things.  GNU make broke things somewhere in version 3 (I forgot). Ocaml has never broken anything, but it started printing warnings in the transition to ocaml4 that people found annoying.  Steven and I are the only ones who need to be aware of these changes, and the "user" doesn't know anything about this.  The "official" fftw tarball works pretty much everywhere, and I do test that it works on as many OS's as I can put my hands on, including all major linux distributions released this century, various *BSD, MacOS, Windows, and (whenever I can) AIX and Solaris using native userland (not GNU).  Yes, autoconf sucks, but one reason for its complexity is that it outputs a script that can only assume very little about its execution environment.

bagong:
Let's not forget though that the source generation part is only half of the PR. The other part is simply a very well done build system for the distributed source that works as well on Windows native as in Linux and MacOS. Cmake *is* developing into the lingua franca of cross platform build scripts.

matteo-frigo:
@bagong You are totally correct: we should distinguish the code-generation process from the build process.  

We are not willing to change our distribution model (e.g., we are not willing to force users to install ocaml and other dependencies in order to regenerate the fftw source tarball), at least for the time being.  That is, "users" will continue to download a tarball from fftw.org and will not have to start from github.  Doing otherwise will break too many things that we are unwilling to break.  We are not willing, e.g., to force users to install ocaml, cmake, or even bash and GNU make, in order to compile fftw.

However, as I said on Jan 19, we are willing to integrate a cmake build infrastructure in parallel with the automake build infrastructure and maintain it going forward, so as to make life easier for users that are using cmake natively.  In this universe, the user downloads the source tarball and then can either run "configure && make" or "cmake".

So here is my offer: If anybody is willing to provide a PR that does that (and only that) and whose resulting CMakefile works on Windows and Linux, we will adopt it and maintain it.  It has to work on Windows otherwise there is no point in doing it.  It has to work on Linux so that I can maintain it.  Of course, the wider the applicability, the better.

jakirkham:
So what about PR ( https://github.com/FFTW/fftw3/pull/103 ) then?

matteo-frigo:
I merged #103

PR #84, opened 2017-01-22
fix __cdecl for pointer return types

EikeVerdenhalven:
Fix cases where functions return pointer types to compile with MSVC


matteo-frigo:
42c0036e should fix the problem in a more disciplined way.  Let me know if it works.

EikeVerdenhalven:
@matteo-frigo : your fix compiles fine here, so I will close this PR

PR #86, opened 2017-01-25
FFTW.NET

ArgusMagnus:
Currently supports N-dimensional complex-to-complex, complex-to-real and real-to-complex transformations.

(This is my first pull request ever, so let me know if there is something missing or wrong with it.)


stevengj:
@ArgusMagnus, thanks for working on this.   For wrappers to call FFTW from other languages, we generally prefer that you just put together a separate package, and then we will be happy to link to it from the "other languages" section of http://fftw.org/download.html

ArgusMagnus:
Ok, thanks for the fast response.
I've created my own repo here: https://github.com/ArgusMagnus/FFTW.NET

stevengj:
Thanks, I've added a link.

PR #107, opened 2017-09-20
Config mode

xantares:
Installs FFTW3Config.cmake instead of a FindFFTW3.cmake
Also configures the pkgconfig file from cmake

follows #69


CMakeLists.txt
```diff
@@ -344,24 +343,53 @@ foreach(subtarget ${subtargets})
...
+    endmacro ()
+
+    fftw_add_test (32x64)
+    fftw_add_test (ib256)
+
+  endif ()
 endif ()
+
+# pkgconfig file
+set (prefix ${CMAKE_INSTALL_PREFIX})
```
junghans: I think, I would be nicer to use the cmake variable (e.g. `CMAKE_INSTALL_PREFIX`) in all the .in files and modify the `Makefile.am` accordingly.

CMakeLists.txt
```diff
@@ -344,24 +343,53 @@ foreach(subtarget ${subtargets})
...
+         DESTINATION ${LIBRARY_PATH}/pkgconfig
+         COMPONENT Development)
+
+# cmake file
+configure_file (FFTW3Config.cmake.in FFTW3${PREC_SUFFIX}Config.cmake @ONLY)
+configure_file (FFTW3ConfigVersion.cmake.in FFTW3${PREC_SUFFIX}ConfigVersion.cmake @ONLY)
+install (FILES
+          ${CMAKE_CURRENT_BINARY_DIR}/FFTW3${PREC_SUFFIX}Config.cmake
+          ${CMAKE_CURRENT_BINARY_DIR}/FFTW3${PREC_SUFFIX}ConfigVersion.cmake
+         DESTINATION ${LIBRARY_PATH}/cmake/fftw3
```
junghans: Instead of `${LIBRARY_PATH}` use `include(GNUInstallDirs)` and `CMAKE_INSTALL_LIBDIR` will have the right value automatically.

Makefile.am
```diff
@@ -142,14 +142,14 @@ fftw3@PREC_SUFFIX@.pc: fftw.pc
...
 	$(SED) \
 		-e 's|[@]PREC_SUFFIX@|@PREC_SUFFIX@|g' \
 		-e 's|[@]libdir@|$(libdir)|g' \
 		-e 's|[@]includedir@|$(includedir)|g' \
-		$(top_srcdir)/FindFFTW3.cmake.in > $@
-cmakedir = $(datarootdir)/cmake/Modules
-cmake_DATA = FindFFTW3@PREC_SUFFIX@.cmake
+		$(top_srcdir)/FFTW3Config.cmake.in > $@
+cmakedir = $(libdir)/cmake/fftw3
+cmake_DATA = FFTW3@PREC_SUFFIX@Config.cmake
```
junghans: I think `FFTW3@PREC_SUFFIX@ConfigVersion.cmake.in` is missing here and the related make rule.

PR #109, opened 2017-09-20
Fix includes, export target

xantares:
I messed includes in the previous PR.
Add target export.


CMakeLists.txt
```diff
@@ -241,8 +241,22 @@ set (fftw_threads_SOURCE ${fftw_par_SOURCE} threads/threads.c)
 set (fftw_omp_SOURCE ${fftw_par_SOURCE} threads/openmp.c)
 
 
-include_directories(.)
-
+include_directories(.
```
matteo-frigo: This should no longer be necessary on master.  See 1f3704b9.  Why do you think the explicit list is necessary?

xantares: You're right

PR #112, opened 2017-10-31
Pr  armv7 pmccntr counter and docs

alexeicolin:
perf counters: add PMCCNTR for ARMv7 and add docs
    
    The existing armv7 counter (CNTVCT) does need enabling from kernel mode (so
    updated the configure help), and the enable bit is different from the PMU
    enable bit (described in the new docs).
    
    Tested on XU4: printed the returned counter values and they look reasonable.

If CNTVCT is not enabled in kernel mode, using the code in the new docs, I get illegal instruction at the instructions that reads CNTVCT.

FWIW, I PRed: https://github.com/thoughtpolice/enable_arm_pmu/pull/6

While at it also did this rename (easy to drop if can't break compatibility with the old --enable-armv8cc):

    perf counters: name ARMv8 PMCCNTR_EL0 explicitly
    
    For consistency with the rest.

RFC: It might be worth it to add the kernel module source into this repo, because those linked repos and forks are scattered and seem like they are not maintained. Integrating into the build is probably out of the question due to the need for kernel sources, but maybe at least install the module source into /usr/share, and read the config vars corresponding to --enable* to include the right instructions into that source. Thoughts?

### DIFF
--- a/README-perfcnt.md
+++ b/README-perfcnt.md
@@ -1,0 +1,93 @@
+Performance Counters
+====================
+
+FFTW measures execution time in the planning stage, optionally taking advantage
+of hardware performance counters. This document describes the supported
+counters and additional steps needed to enable each on different architectures.
+
+See `./configure --help` for flags for enabling each supported counter.
+See [kernel/cycle.h](kernel/cycle.h) for the code that accesses the counters.
+
+ARMv7-A (armv7a)
+================
+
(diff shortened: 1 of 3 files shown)


matteo-frigo:
Pulled, thanks.

What's your real name?  README-perfcnt should list you as an author.

PR #119, opened 2017-11-25
Exports some internal symbols so that we won't have to WITH_COMBINED_THREADS for Win32.

sxprophet:
`X(mksolver_ct_hook)` and `X(mksolver_hc2hc_hook)` are variables in which case we have to explicitly modify them with `__declspec(dllimport)` in the headers.

### DIFF
--- a/api/apiplan.c
+++ b/api/apiplan.c
@@ -22,7 +22,7 @@
 
 static planner_hook_t before_planner_hook = 0, after_planner_hook = 0;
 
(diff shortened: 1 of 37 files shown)


jimwang118:
What is the purpose of adding this code in ct.h?
```
#if (defined(FFTW_DLL) || defined(DLL_EXPORT)) && (defined(_WIN32) || defined(__WIN32__))
  __declspec(dllimport)
#endif
```

PR #121, opened 2017-12-22
Fix deadlock on Windows with shared libraries

SimonRit:
I have faced deadlocks on Windows with FFTW DLLs. This is described in details on [ITK discourse](https://discourse.itk.org/t/deadlock-with-fftw-shared-libs-and-windows/494). The proposed patch fixes the problem. I'm happy to do another test if you find a neater implementation for the bug fix.

### DIFF
--- a/threads/threads.c
+++ b/threads/threads.c
@@ -156,6 +156,8 @@ typedef pthread_mutex_t os_static_mutex_t;
 static void os_static_mutex_lock(os_static_mutex_t *s) { pthread_mutex_lock(s); }
 static void os_static_mutex_unlock(os_static_mutex_t *s) { pthread_mutex_unlock(s); }
 
+#define NOT_SIGNALED 1
+
 #elif defined(__WIN32__) || defined(_WIN32) || defined(_WINDOWS)
 /* hack: windef.h defines INT for its own purposes and this causes
    a conflict with our own INT in ifftw.h.  Divert the windows
@@ -245,6 +247,9 @@ static void os_static_mutex_unlock(os_static_mutex_t *s)
      LONG old = InterlockedExchange(s, 0);
      A(old == 1);
 }
+
+#define NOT_SIGNALED (WaitForSingleObject(q->ready, 0) == WAIT_TIMEOUT)
+
 #else
 #error "No threading layer defined"
 #endif
@@ -367,7 +372,8 @@ static void kill_workforce(void)
 	       worker_queue = q->cdr;
 	       q->w = &w;
 	       os_sem_up(&q->ready);
-	       os_sem_down(&termination_semaphore);
+	       if (NOT_SIGNALED)
+	            os_sem_down(&termination_semaphore);
 	       unmake_worker(q);
 	  }
      });


SimonRit:
@dzenanz The bug does not occur if you use FFTW in form of DLL on Windows and call the cleanup functions before the main function ends. It occurs in ITK because the FFTW implementation there is based on a singleton which destroys itself after leaving the main function.

matteo-frigo:
This "fix" is incorrect and works by accident.  With the proposed fix, kill_workforce executes sem_up(&ready) followed by an immediate sem_down(&ready) in WaitForSingleObject(&ready, 0).  This breaks the termination prototol and leaks worker threads.  Of course the fix appears to work: it is equivalent to not calling kill_workforce() at all.  The problem is that the "fix" also leaks threads in all the other cases.

Can anybody explain exactly what the problem is and why you think it is our problem?  My understanding is that you are initializing FFTW as a "singleton object" outside of main.  You call kill_workforce() in the "destructor" of the singleton object.  However, windows terminates threads when main() exists, and now the destructor hangs.  If that's the case, I don't see how this has anything to do with FFTW.  Obviously windows is broken.  If you insist that windows is not broken, obviously this "singleton object" thing is broken for any "object" that contains threads, so you shouldn't use it in the first place.  If you insist on using it, don't call kill_workforce() at all from the "destructor" and live with the memory leak.

SimonRit:
Hi,
I generally agree although I don't understand what you mean by all other cases? The "fix" should not change the behavior if the threads are still running.
I don't know if Windows is broken or if there is something that must be done differently, I couldn't figure it out better than I did here. I have implemented an [even dirtier patch in ITK](https://github.com/InsightSoftwareConsortium/ITK/commit/4123647084aa94b7377ff625dba366a0df498518) because I was already guessing your reluctance... I'll refer to our discussion to the ITK thread because I'm guessing some developers there can explain.
Simon

matteo-frigo:
The fix will change the behavior if the threads are sill running, that's exactly the problem.

Your fix, after macro expansion, says
 
   sem_up(&ready);
   sem_down(&ready);  /* implemented as WaitForSingleObject with timeout */
   if (...) sem_down(&terminate);

So kill_workforce() consumes the sem_up(&ready) that was meant for the worker threads.  Thus, the worker threads will not terminate properly.

threads/threads.c
```diff
@@ -245,6 +247,9 @@ static void os_static_mutex_unlock(os_static_mutex_t *s)
      LONG old = InterlockedExchange(s, 0);
      A(old == 1);
 }
+
+#define NOT_SIGNALED (WaitForSingleObject(q->ready, 0) == WAIT_TIMEOUT)
```
dzenanz: Perhaps you should try waiting for a thread here, not the semaphore, something like this: `#define NOT_SIGNALED (WaitForSingleObject(m_Threads[0], 1) == WAIT_OBJECT_0)`

SimonRit:
@matteo-frigo I don't think my code looks like that after macro expansion. At least it should be
```
if (WaitForSingleObject(q->ready, 0) == WAIT_TIMEOUT) os_sem_down(&termination_semaphore);
```
which I have written based on the [doc of WaitForSingleObject](https://msdn.microsoft.com/en-us/library/windows/desktop/ms687032(v=vs.85).aspx):
> If dwMilliseconds is zero, the function does not enter a wait state if the object is not signaled; it always returns immediately.
That's the best I could find out to test if q->ready had been signaled.
@dzenanz I don't know what's best but between testing the thread or the mutex. Can you explain your suggestion? Also, why 1 and not 0 for the timeout?

threads/threads.c
SimonRit: @dzenanz I don't know what's best but between testing the thread or the mutex. Can you explain your suggestion? Also, why 1 and not 0 for the timeout?

matteo-frigo:
@SimonRit Ok, let me try again more explicitly.
The line before your fix increments the semaphore q->ready.  The worker thread is supposed to wait on that semaphore.
Your fix, specifically "WaitForSingleObject(q->ready, 0)", behaves in one of two ways.  If the main thread executes WaitForSingleObject(q->ready, 0) before the worker thread, then the main thread decrements the semaphore q->ready and the worker thread never terminates.  This is bad.  If the worker thread gets there first, then WaitForSingleObject(q->ready, 0) returns WAIT_TIMEOUT and nothing bad happens.

The problem is that either case is possible, and the first alternative breaks the termination protocol.  For example, a program that repeatedly initializes and un-initializes fftw will end up leaking threads.

To be honest, I would be really surprised if there were a correct way to accomplish the behavior that you are looking for.  I don't think it is possible *at all*.

dzenanz:
I implemented a [solution](https://github.com/InsightSoftwareConsortium/ITK/commit/25a149ff6ceffcaa8304f2338f49d904f4efce0d#diff-595b2adcc597f969ff47a6ec080c783eL185) for just such a problem in ITK.

`#if  Win32+DLL` check whether the thread has already exited, and if so skip thread termination protocol via semaphores. Otherwise waiting on a semaphore which the thread is supposed to signal causes a deadlock.

SimonRit:
Ok, got it, thanks for taking the time to explain. Since the thread handle (output of _beginthreadex) is not stored, I cannot easily implement the suggestion of @dzenanz and I don't see any simple solution.

Issue #122, opened 2018-01-05
fftw hangs on raspberry pi3 


soundart:
Currently fftw does not work on my raspberry pi3. It hangs in a loop doing calls to clock_gettime:

    clock_gettime(CLOCK_SGI_CYCLE, 0x7ea42558) = -1 EINVAL (Invalid argument)
    clock_gettime(CLOCK_SGI_CYCLE, 0x7ea42558) = -1 EINVAL (Invalid argument)
    clock_gettime(CLOCK_REALTIME, {tv_sec=1513978172, tv_nsec=339522337}) = 0
    clock_gettime(CLOCK_REALTIME, {tv_sec=1513978172, tv_nsec=339672025}) = 0
    clock_gettime(CLOCK_SGI_CYCLE, 0x7ea42558) = -1 EINVAL (Invalid argument)

The call clock_gettime(CLOCK_SGI_CYCLE, &t) from function getticks() fails with EINVAL.
See : http://elixir.free-electrons.com/linux/latest/source/include/uapi/linux/time.h#L62 - "The driver implementing this got removed. The clock ID is kept as a place holder. Do not reuse!"

There was a commit preventing the usage of CLOCK_SGI_CYCLE on ANDROID. It should probably not be used on plain linux too.

I discussed this here: https://github.com/FFTW/fftw3/commit/950b153910f7f0dde9cc20cddeee5dc9048d25b7 

I will add a pull request based on https://github.com/SETIguy comment https://github.com/FFTW/fftw3/commit/950b153910f7f0dde9cc20cddeee5dc9048d25b7#commitcomment-26583436

soundart:
See https://github.com/FFTW/fftw3/pull/123

paigeadelethompson:
Yeah Im having this issue too now myself, how is it that I am the only person who is though ???? This should be a pretty big deal by my account (cant start either gqrx nor sdrangel) Am I understanding correctly that nobody actually uses these on Raspi ???? 

```
#if defined(HAVE_CLOCK_GETTIME) && defined(CLOCK_SGI_CYCLE) && !defined(HAVE_TICK_COUNTER) && !defined(__ANDROID__)
typedef struct timespec ticks;

static inline ticks getticks(void)
{
     struct timespec t;
     clock_gettime(CLOCK_SGI_CYCLE, &t);
     return t;
}

```

PR #123, opened 2018-01-05
linux: avoid using CLOCK_SGI_CYCLE

soundart:
The call clock_gettime(CLOCK_SGI_CYCLE, &t) from function getticks() fails with EINVAL.
See : http://elixir.free-electrons.com/linux/latest/source/include/uapi/linux/time.h#L62 - "The driver implementing this got removed. The clock ID is kept as a place holder. Do not reuse!"

See https://github.com/FFTW/fftw3/issues/122

### DIFF
--- a/kernel/cycle.h
+++ b/kernel/cycle.h
@@ -437,8 +437,18 @@ INLINE_ELAPSED(__inline)
 #define HAVE_TICK_COUNTER
 #endif
 /*----------------------------------------------------------------*/
-/* SGI/Irix */
-#if defined(HAVE_CLOCK_GETTIME) && defined(CLOCK_SGI_CYCLE) && !defined(HAVE_TICK_COUNTER) && !defined(__ANDROID__)
+/* SGI/Irix/Linux but not android */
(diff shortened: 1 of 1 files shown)


soundart:
Hi I updated the patch, because:
- I forgot to use the CLOCK_XXX/METHOD parameter
- It made problems on linux amd64 because there the X86-64 cycle counter was used with a conflicting definition of ticks

soundart:
Hi,

yes I think too, but it is difficult to judge without an android build for testing. 

I guess the android build uses one of these:

#if defined(HAVE_ARMV7A_CNTVCT)
or
#if defined(HAVE_ARMV7A_PMCCNTR)
or
#if defined(__aarch64__) && defined(HAVE_ARMV8_CNTVCT_EL0) && !defined(HAVE_ARMV8_PMCCNTR_EL0)
or
#if defined(__aarch64__) && defined(HAVE_ARMV8_PMCCNTR_EL0)


 and I think these defines should move upwards in cycle.h, so that they appear before the more generic clock_gettime()

Issue #133, opened 2018-03-02
a few comments about windows build and the instruction page


HinTak:
about http://www.fftw.org/install/windows.html .

It is possible to create visual studio import libraries without using visual studio tool `lib.exe`, by using the mingw dlltool as follows:
```
i686-w64-mingw32-dlltool -l libfftw3-3.lib -d  libfftw3-3.def
x86_64-w64-mingw32-dlltool -l libfftw3-3.lib -d  libfftw3-3.def
```

It would be nice if this is integrated with the cross build scripts.

Under section "Compiling FFTW 3 for Windows" subsection "Building FFTW 3 under MinGW", ```--with-our-malloc16``` is for older fftw than current. It seems that in current code, this configure option has been renamed to just ```--with-our-malloc``` without the 16. Would be nice to include both of these in that description and say one is current and older fftw uses the other.

stevengj:
`--with-our-malloc16` actually [still works](https://github.com/FFTW/fftw3/blob/eba07c46b5d2f7824d293ab59aa5c29a25034963/configure.ac#L257), so we can use it to support both newer and older versions if needed.

HinTak:
Another issue worths mentioning is that "Image has Safe Exception Handlers" 
(https://docs.microsoft.com/en-gb/cpp/build/reference/safeseh-image-has-safe-exception-handlers) is AFAIC on by default, and trying to link an application with gcc-built dlls in visual studio fails because mingw gcc does not currently supports SEFESEH . So one needs to switch this off. This option is under the "Advanced" tab of the linker property section in visual studio.

There is nothing one can do other than switching it off at the visual studio end (and taking whatever drawback switching it off brings).

HinTak:
Another thing worth mentioning is that mingw-built 32-bit fftw dll can have a dependency on libgcc_s_sjlj-1.dll ( and in turn libwinpthread-1.dll ). 64-bit fftw dll does not have this issue. What this means is that if that's the case, it would be advisable or even necessary to bundle the matching versions of thoese two dlls in a binary distribution.

I am not aware of a better and more stand-alone tool for examining dll dependencies in a cross-compiling environment - I myself use `winedump -j import ...` for that purpose, though it is perhaps considered overkill.

HinTak:
I was just commenting about the lack of "official" Microsoft tool chain support, but Microsoft provides one, via cmake, interestingly enough.

https://github.com/Microsoft/vcpkg/tree/master/ports/fftw3

[...]

PR #135, opened 2018-03-07
Uses dlltool from GNU binutils to generate VC import libraries, and u…

HinTak:
…pdate README accordingly.

zipping up the resulting import libraries also.

This addresses some of the issues raised in https://github.com/FFTW/fftw3/issues/133 .


HinTak:
Tested on WSL Ubuntu under windows 10.

HinTak:
Ping - it has been a while!

HinTak:
Thanks - it has been 3 years!

PR #141, opened 2018-04-10
Define include directory for installed targets

### LINKED ISSUE #140: Include directory not defined for installed targets
FFTW can be incorporated into other projects using a FindPackage command, thanks to the FFTWConfig.cmake file that is provided at the install.

However, the include directories are not set for the targets, and to get this work, one needs to add the "include" subdirectory of the FFTW install prefix to the include directories of their project manually. This should not be necessary if the target_include_directories property would be set for the installed targets.

espakm:
Fixes #140.

Signed-off-by: Miklos Espak <espakm@gmail.com>


CMakeLists.txt
```diff
@@ -304,6 +304,7 @@ if (BUILD_SHARED_LIBS)
 endif ()
 
 add_library (${fftw3_lib} ${SOURCEFILES})
+target_include_directories (${fftw3_lib} INTERFACE $<INSTALL_INTERFACE:include>)
```
stevengj: Should there also be a `$<BUILD_INTERFACE:api>`?

espakm: It would make sense, but as I found, it is not needed. I tried now to use FFTW from the build folder instead of the install folder and the code compiled even without `${BUILD_INTERFACE:api}`.

PR #150, opened 2018-07-31
Switching on CNTVCT for ThunderX2 platform and setting some optimization options.

aur-ml:
Tested on ThunderX2 servers: the performance on 2^20 i-FFT problem (ib1048576) improves for about 20-25%.
The proposed changes does not influence other hardware architectures.


m4/ax_cc_maxopt.m4
```diff
@@ -99,6 +99,23 @@ if test "$ac_test_CFLAGS" != "set"; then
...
+         AX_CHECK_COMPILER_FLAGS(-fdevirtualize, CFLAGS="$CFLAGS -fdevirtualize")
+         AX_CHECK_COMPILER_FLAGS(-fearly-inlining, CFLAGS="$CFLAGS -fearly-inlining")
+         AX_CHECK_COMPILER_FLAGS(-ffunction-cse, CFLAGS="$CFLAGS -ffunction-cse")
+         AX_CHECK_COMPILER_FLAGS(-fgcse-after-reload, CFLAGS="$CFLAGS -fgcse-after-reload")
+         AX_CHECK_COMPILER_FLAGS(-fforward-propagate, CFLAGS="$CFLAGS -fforward-propagate")
+         AX_CHECK_COMPILER_FLAGS(-fpack-struct, CFLAGS="$CFLAGS -fpack-struct")
+         AX_CHECK_COMPILER_FLAGS(-fsched-critical-path-heuristic, CFLAGS="$CFLAGS -fsched-critical-path-heuristic")
+         AX_CHECK_COMPILER_FLAGS(-fsel-sched-pipelining, CFLAGS="$CFLAGS -fsel-sched-pipelining")
+         AX_CHECK_COMPILER_FLAGS(-fselective-scheduling, CFLAGS="$CFLAGS -fselective-scheduling")
+         AX_CHECK_COMPILER_FLAGS(-ftrapping-math, CFLAGS="$CFLAGS -ftrapping-math")
```
stevengj: Are all of these flags necessary for the performance improvements, or is it only one or two of them that really matters?

aur-ml: > Are all of these flags necessary for the performance improvements, or is it only one or two of them that really matters?

Among about a hundred of checked flags only those were left, that have statistically significant influence on performance for the target platform.

aur-ml: Here are charts with several benchFFT tests for optimized vs default versions:
https://github.com/aur-ml/fftw3/wiki

ashwinyes:
@stevengj Request you to please review the code.

Please merge if the the PR looks OK. Else please let @aur-ml know if there are any changes that needs to be made.

Thank you

rdolbeau:
This is very old, and I'm not sure a bunch of flags that work on a specific micro-architecture belongs in configure. Those flags can be provided to the build system from the command line anyway.

PR #154, opened 2018-09-26
Fix compiling omp tests

jtravs:


ZedThree:
This also needs `target_link_libraries (${fftw3_lib}_omp OpenMP::OpenMP_C)` so that the OpenMP dependency is propagated to targets.

CMakeLists.txt
```diff
@@ -385,6 +385,11 @@ if (BUILD_TESTS)
     target_link_libraries (bench ${fftw3_lib})
   endif ()
 
+  if (ENABLE_OPENMP)
+    target_link_libraries (bench ${fftw3_lib}_omp)
+  else ()
+    target_link_libraries (bench ${fftw3_lib})
```
stevengj: I don't think this `else` clause is needed, since `${fftw3_lib}` should have already been added by the preceding `if/else`.

PR #155, opened 2018-10-07
Fix typo in Fortran quadruple-precision doc.

jme52:
The use of quadruple-precision with Fortran does not require the inclusion of file `fftw3l.f03` (which contains the interfaces for extended precision), but it requires the inclusion of file `fftw3.f03`.


stevengj:
Thanks.

PR #191, opened 2020-03-11
Small performance improvement for avx-512 on Skylake-SP

rdolbeau:
This improves performance a bit on the Skylake-SP cores (Xeon Scalable), by replacing gather/scatter by slightly more efficient code: breaking down the instruction in 128 bits chunk for DP, and going for 64-bits scatter/gather (instead of 32) in SP. The original gather/scatter code is still available for DP, as it's probably faster on Knights Landing (KNL, Xeon Phi 72xx). The SP code should be a win on KNL as well.

Tested with make check/bigcheck, and for performance on synthetic code, could probably use some real-life testing for performance.

### DIFF
--- a/simd-support/simd-avx512.h
+++ b/simd-support/simd-avx512.h
@@ -99,31 +99,19 @@ static inline void STA(R *x, V v, INT ovs, const R *aligned_like) {
 static inline V LDu(const R *x, INT ivs, const R *aligned_like)
 {
   (void)aligned_like; /* UNUSED */
-  __m512i index = _mm512_set_epi32(7 * ivs + 1, 7 * ivs,
-                                   6 * ivs + 1, 6 * ivs,
-                                   5 * ivs + 1, 5 * ivs,
-                                   4 * ivs + 1, 4 * ivs,
-                                   3 * ivs + 1, 3 * ivs,
-                                   2 * ivs + 1, 2 * ivs,
-                                   1 * ivs + 1, 1 * ivs,
-                                   0 * ivs + 1, 0 * ivs);
+  /* pretend pair of single are a double */
+  const __m256i index = _mm256_set_epi32(7 * ivs, 6 * ivs, 5 * ivs, 4 * ivs, 3 * ivs, 2 * ivs, 1 * ivs, 0 * ivs);
   
-  return _mm512_i32gather_ps(index, x, 4);
+  return (V)_mm512_i32gather_pd(index, x, 4);
 }
 
 static inline void STu(R *x, V v, INT ovs, const R *aligned_like)
 {
   (void)aligned_like; /* UNUSED */
-  __m512i index = _mm512_set_epi32(7 * ovs + 1, 7 * ovs,
-                                   6 * ovs + 1, 6 * ovs,
-                                   5 * ovs + 1, 5 * ovs,
-                                   4 * ovs + 1, 4 * ovs,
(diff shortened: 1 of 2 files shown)


simd-support/simd-avx512.h
```diff
@@ -99,55 +99,63 @@ static inline void STA(R *x, V v, INT ovs, const R *aligned_like) {
...
-                                   4 * ivs + 1, 4 * ivs,
-                                   3 * ivs + 1, 3 * ivs,
-                                   2 * ivs + 1, 2 * ivs,
-                                   1 * ivs + 1, 1 * ivs,
-                                   0 * ivs + 1, 0 * ivs);
+  /* pretend pair of single are a double */
+  const __m256i index = _mm256_set_epi32(7 * ivs, 6 * ivs, 5 * ivs, 4 * ivs, 3 * ivs, 2 * ivs, 1 * ivs, 0 * ivs);
   
-  return _mm512_i32gather_ps(index, x, 4);
+  return (V)_mm512_i32gather_pd(index, x, 4);
```
xiegengxin, line 105: return (V)_mm512_i32gather_pd(index, x, 8);
Right?

simd-support/simd-avx512.h
```diff
@@ -99,55 +99,63 @@ static inline void STA(R *x, V v, INT ovs, const R *aligned_like) {
...
-                                   3 * ovs + 1, 3 * ovs,
-                                   2 * ovs + 1, 2 * ovs,
-                                   1 * ovs + 1, 1 * ovs,
-                                   0 * ovs + 1, 0 * ovs);
-  
-  _mm512_i32scatter_ps(x, index, v, 4);
+  /* pretend pair of single are a double */
+  const __m256i index = _mm256_set_epi32(7 * ovs, 6 * ovs, 5 * ovs, 4 * ovs, 3 * ovs, 2 * ovs, 1 * ovs, 0 * ovs);
+
+  _mm512_i32scatter_pd(x, index, (__m512d)v, 4);
```
xiegengxin, line 114: _mm512_i32scatter_pd(x, index, (__m512d)v, 8);

xiegengxin (changes requested):
I think the scale should be 8 when using Scatter/Gather to access 2 32-bit elements as one 64-bit.

simd-support/simd-avx512.h
rdolbeau: Wrote that code a while ago, but I think 4 is correct; the indices are still referring to the original datatype - single precision value of 4 bytes. The`_pd` variant is used only to access 64 bits at a time explicitly.

simd-support/simd-avx512.h
rdolbeau: Same here - `ovs` is a stride in 4-bytes elements, so the `index` vector is also in 4-bytes element.

simd-support/simd-avx512.h
xiegengxin: You're right.

rdolbeau:
I've updated the commit message after (finally) testing on KNL.

rdolbeau:
@stevengj @matteo-frigo Can I merge this (old) one? I don't think KNL not liking the new code will be much of a problem by now, as I think most of the KNL-based systems have been retired (and there's a configure option to produce a KNL-friendly version anyway, as I still do own a KNL myself :-) )

PR #205, opened 2020-06-19
First attempt at adding planner_nthreads

galenlynch:
While you can currently set the maximum number of threads used by the planner,
there is no way to check that number. This simply adds a new function,
`planner_nthreads`, that provides this information, as was suggested by
@stevengj (JuliaMath/FFTW.jl#150).

I have no idea which files I should edit to introduce this new function beyond `threads/api.c`, due to my unfamiliarity with autoconf and the build system used by FFTW. I have additionally added `planner_nthreads` to `api/fftw3.h`, and `threads/f77funcs.h`. I was unsure how to test this new function, and I have not yet done so. I will try to figure that out soon, if someone with more experience does not do it before I can. 

I have also added `planner_nthreads` to all the relevant portions of the documentation that I could find.

### DIFF
--- a/api/fftw3.h
+++ b/api/fftw3.h
@@ -380,6 +380,9 @@ FFTW_EXTERN void                                                        \
 FFTW_CDECL X(plan_with_nthreads)(int nthreads);                         \
                                                                         \
 FFTW_EXTERN int                                                         \
+FFTW_CDECL X(planner_nthreads)(void);                                   \
+                                                                        \
+FFTW_EXTERN int                                                         \
 FFTW_CDECL X(init_threads)(void);                                       \
                                                                         \
 FFTW_EXTERN void                                                        \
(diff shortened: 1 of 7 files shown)


galenlynch:
I successfully built this branch, and added a test of `planner_nthreads` into `fftw-bench.c` that confirms that `planner_nthreads` returns the last argument passed to `plan_with_nthreads`. After adding this, `make check` runs without error.

I'm not sure if this is the best place to test this new function, or if there are other tests that should also be added. If so, I would be happy to add them to this PR.

stevengj:
LGTM.   Probably you could add a test [after this line](https://github.com/FFTW/fftw3/blob/d832ca0e2f286650278563c6887c4f235785e606/tests/fftw-bench.c#L127), of the form:
```c
BENCH_ASSERT(FFTW(planner_nthreads)() == nthreads);
```
There is also some copyright-assignment paperwork that MIT requires for contributions to FFTW; @matteo-frigo, can you point @galenlynch in the right direction for this?

galenlynch:
> ```c
> BENCH_ASSERT(FFTW(planner_nthreads)() == nthreads);
> ```

I think that's the exact test that I added in the second commit of this PR, which is great! Happy to do whatever paperwork is required. I'm an MIT employee if that makes a difference.

galenlynch:
Is this ok to merge?

PR #232, opened 2021-02-24
Fix SIMD detection for MSVC

ruilvo:
This PR fixes a CMake quirk where one needs to delete the loop variable both in cache and locally for it to work. I couldn't compile with SIMD on MSVC without this fix.


stevengj:
Thanks!

PR #241, opened 2021-05-13
Remove `-march=core-avx2`

SoapGentoo:
* On modern compilers, when compiling with

    CFLAGS="-O2 -pipe -march=x86-64"

  and passing `--enable-avx2`, the user's `-march=` will override
  the `-march=core-avx2` from configure.ac, implicitly disabling AVX2
  support again. The fundamental problem is that -march options aren't
  stackable, and configure.ac should **never** add a -march flag, as the
  user should always be allowed to override them. At this point, GCC 4.8
  is obsolete.

Bug: https://bugs.gentoo.org/698572

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -359,12 +359,9 @@ case "${ax_cv_c_compiler_vendor}" in
         fi
 
         # AVX2
-        # gcc-4.8 works with -march=core-avx2, but -mavx2 is not enough.
-        # Later versions seem to happy with -mavx2, so try the arch one first.
         if test "$have_avx2" = "yes" -a "x$AVX2_CFLAGS" = x; then
-            AX_CHECK_COMPILER_FLAGS(-march=core-avx2, [AVX2_CFLAGS="-march=core-avx2"],
-                [AX_CHECK_COMPILER_FLAGS(-mavx2, [AVX2_CFLAGS="-mavx2"],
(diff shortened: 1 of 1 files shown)


SoapGentoo:
@matteo-frigo any chance of getting this merged?

stevengj:
LGTM.  I agree that it shouldn't be a problem to drop gcc 4.8 at this point.

matteo-frigo:
gcc-4.8 is standard in redhat/centos/oraclelinux 7 which are widely used.  No way we can drop it.

SoapGentoo:
can we at least switch the precedence order? Trying the `-march` first kills lots of consumers in Gentoo

matteo-frigo:
Fixed in 62e605c8  (sorry I didn't notice that you actually had a pull request, so I rewrote exactly the same code with a longer commit message).

It seems that the old comment about gcc-4.8 is not true, or at least no longer true.  gcc-4.8 on both CentOS7 and Oracle Linux 7 works fine without -march=..., and so does gcc-4.8 on 32-bit debian jessie (I was worried that maybe things work fine on x86_64 but not on i686, but i686 -mavx2 works fine too).

So I think we are good.  I won't merge your PR, but equivalent code is now on master.

matteo-frigo:
By the way, for gentoo users hit by this, one workaround is to call

  configure ... AVX2_CFLAGS='-mavx2'

which disables the autodetection of AVX2_CFLAGS

PR #296, opened 2022-09-24
Fix mistakes

ivan-pan:
I fixed a few mistakes in the FFTW reference (note that _et cetera_ is written in two words; I used _etc._ instead).


stevengj:
LGTM, thanks!

PR #314, opened 2023-03-02
add fftw_copy_plan

### LINKED ISSUE #106: Proposing `fftw_clone()` to allow memory-safe C++ wrappers
One problem when using FFTW in non-trivial C++ applications is memory safety.  This is particularly the case if an FFTW plan is embedded as a field in a class that should be copy-constructible and assignable:
```C++
struct X {
   X() { fftw_plan_many_dft( /* ... */ ); }
   ~X() { fftw_destroy(plan); /* freeing arrays etc. */ }
   X(const X &other) { /* what do we put here??? */ }
private:
   fftw_plan plan;
   double *in, *out;
};
```
As far as I can tell, there is no way to *clone*, i.e. to obtain a deep copy, of a plan, which is a problem in the copy constructor (and the assignment operator).  An immediate workaround is to re-create the plan in the new object and copy all the data from the arrays.  However, since `fftw_plan` is an opaque pointer, there is no portable way to recreate the plan from itself, which means the above snippet has to re-implemented for every use case.

I therefore propose to add the following function (or something similar) to the public API:
```C
/** Return an exact deep copy (clone) of the plan, operating on new fields
 * 
 * Expects:
 *  - blueprint: original plan to be copied
 *  - new_in, new_out: the cloned plan will operate on these fields.  The data of the
 *    original plan is not copied to these fields.  The same restrictions as for 
 *    `fftw_execute_dft_*` apply.  `fftw_complex` fields should be cast to `double*`
 */
fftw_plan fftw_clone(const fftw_plan blueprint, double *new_in, double *new_out);
```

This would immediately allow to write a copy-constructable wrapper class around a generic `fftw_plan`.

If this proposal is accepted, I can help with the implementation, but I would need some pointers :)

stevengj:
Closes #106.

Since plan execution is read-only, it seemed sufficient to implement this simply by adding an internal reference count to `fftw_plan`, so that `fftw_plan` just increments the counter and `fftw_destroy` plan decrements the counter, only de-allocating the plan when the reference count drops to zero.

@matteo-frigo, what do you think?

### DIFF
--- a/api/api.h
+++ b/api/api.h
@@ -62,6 +62,7 @@ extern "C"
 struct X(plan_s) {
      plan *pln;
      problem *prb;
+     size_t refcount;
      int sign;
 };
 
@@ -107,7 +108,7 @@ apiplan *X(mkapiplan)(int sign, unsigned flags, problem *prb);
 rdft_kind *X(map_r2r_kind)(int rank, const X(r2r_kind) * kind);
 
 typedef void (*planner_hook_t)(void);
-                                                     
+
 void X(set_planner_hooks)(planner_hook_t before, planner_hook_t after);
 
 #ifdef __cplusplus
--- a/api/apiplan.c
+++ b/api/apiplan.c
@@ -52,7 +52,7 @@ static plan *mkplan(planner *plnr, unsigned flags,
 		    const problem *prb, unsigned hash_info)
 {
      plan *pln;
-     
+
      pln = mkplan0(plnr, flags, prb, hash_info, WISDOM_NORMAL);
 
      if (plnr->wisdom_state == WISDOM_NORMAL && !pln) {
@@ -93,10 +93,10 @@ apiplan *X(mkapiplan)(int sign, unsigned flags, problem *prb)
                                          FFTW_PATIENT, FFTW_EXHAUSTIVE};
      int pat, pat_max;
      double pcost = 0;
-     
+
      if (before_planner_hook)
           before_planner_hook();
-     
+
      plnr = X(the_planner)();
 
      if (flags & FFTW_WISDOM_ONLY) {
@@ -140,6 +140,7 @@ apiplan *X(mkapiplan)(int sign, unsigned flags, problem *prb)
 	  /* build apiplan */
 	  p = (apiplan *) MALLOC(sizeof(apiplan), PLANS);
 	  p->prb = prb;
+	  p->refcount = 1u;
 	  p->sign = sign; /* cache for execute_dft */
 
 	  /* re-create plan from wisdom, adding blessing */
@@ -172,8 +173,22 @@ apiplan *X(mkapiplan)(int sign, unsigned flags, problem *prb)
 
      if (after_planner_hook)
           after_planner_hook();
-     
+
      return p;
+}
+
+X(plan) X(copy_plan)(X(plan) p)
+{
+     if (p) {
+          if (before_planner_hook)
+               before_planner_hook();
+
+          p->refcount++;
+
+          if (after_planner_hook)
+               after_planner_hook();
+     }
+	 return p;
 }
 
 void X(destroy_plan)(X(plan) p)
@@ -181,11 +196,13 @@ void X(destroy_plan)(X(plan) p)
      if (p) {
           if (before_planner_hook)
                before_planner_hook();
-     
-          X(plan_awake)(p->pln, SLEEPY);
-          X(plan_destroy_internal)(p->pln);
-          X(problem_destroy)(p->prb);
-          X(ifree)(p);
+
+          if (p->refcount-- == 1u) {
+               X(plan_awake)(p->pln, SLEEPY);
+               X(plan_destroy_internal)(p->pln);
+               X(problem_destroy)(p->prb);
+               X(ifree)(p);
+          }
 
           if (after_planner_hook)
                after_planner_hook();
--- a/api/apiplan.c
+++ b/api/apiplan.c
@@ -188,7 +188,7 @@ X(plan) X(copy_plan)(X(plan) p)
           if (after_planner_hook)
                after_planner_hook();
      }
-	 return p;
+     return p;
 }
 
 void X(destroy_plan)(X(plan) p)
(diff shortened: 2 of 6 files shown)


matteo-frigo:
My first reaction is to say that this is a bad idea.

If the use case is C++ wrappers, C++ already has standard facilities for wrapping plain old pointers into reference-counted objects, with a menu of mutually exclusive semantics (shared vs unique vs move).  I don't see what problem is solved by reimplementing this functionality at a lower level.

Is there some case where std::make_shared doesn't do what you need?

Or perhaps you have some other language in mind?

stevengj:
I was actually thinking of Julia and Python, if they want to implement a `deepcopy` method.   Of course, in those languages one can (and does) already wrap an `fftw_plan` in a garbage-collected object, which often contains some other information as well (e.g. about the size and type of array that the plan is applicable to, since we don't have a way to extract this from the `fftw_plan`).  If for some reason the user requests a `deepcopy` of this object you are supposed to return a new high-level object, but since you can't copy the `fftw_plan` you'd need to either forbid `deepcopy` or wrap the plan in another layer of garbage collection.    It's not a huge deal, just a slight annoyance that there is no way to copy a plan.

And of course it might be useful in C and low-level languages as well, which don't have built-in reference counting or garbage collection, if you want to pass plan pointers around without having to keep track of who owns the pointer.

matteo-frigo:
I don't know, it still smells like a bad idea to me, but if you think this is needed, go ahead and push it.

My main objection is that this change doesn't solve any real problems, since I am afraid that every real use case will end up wrapping a plan into some other high-level object, at which point you may as well use native facilities to resource-manage the object.  You allude to this problem yourself when you say that the object must contain some other information anyway, e.g. array size.

Your comment on make_shared is valid.  I had in mind the case where the plan is wrapped in a C++ object, because no self-respecting C++ programmer would be caught dead using a raw C pointer.

But I don't feel strongly about this issue.  If you push it, maybe add a test somewhere.

stevengj:
There is already a test

stevengj:
We should probably set up GitHub CI at some point…

stevengj:
I'm going to go ahead and merge.

correaa:
I also badly need this; I hope it becomes mainstream (e.g., in other implementations that use the FFTW interface).

What critics of this idea miss is that having independent copies of the plans is not about multiple objects that share a reference;
but it is about multiple equivalent plans that can be used simultaneously from different threads without internal locks.
(So, allocated buffers (or any internal state) are not shared.)

And, of course, one wouldn't want to repeat the estimate/measurement computations even if one can preserve the input/output layouts.

I always assumed this would be eventually implemented with a `fftw_sscan_plan`, analogous to `fftw_sprint_plan`.

matteo-frigo:
@correaa fftw plans are already "thread safe" in the sense that the plan is read-only when executing the plan, and thus you can execute the same plan from multiple threads.  Plans do not store buffers, but they do store immutable arrays of trigonometric constants and other read-only information.  So I don't think that the problem you mention needs to be solved.

correaa:
> @correaa fftw plans are already "thread-safe" in the sense that the plan is read-only when executing the plan, and thus you can execute the same plan from multiple threads.  Plans do not store buffers, but they do store immutable arrays of trigonometric constants and other read-only information.  So I don't think that the problem you mention needs to be solved.

OK, I have been awfully mistaken for a while then.

I was convinced that plan execution (over disjoint targets) was not thread-safe. In my mind, it made sense because I assumed that the plan creation would allocate buffers to perform the ffts, so executions themselves wouldn't need to do it at that point. 

I was so convinced about this that I misread the documentation:
http://www.fftw.org/fftw3_doc/Thread-safety.html, which was silly on my side. I confused the comment about planner (plan creation) with the plan execution itself. 

In that case, I wonder if there are buffers behind the scenes in some global state of the library. 
And if so, how they are synchronized?
(I can be mistaken again but it looks like since the planning is not thread safe then there is a lot of global state behind the scenes.)

Anyhow, given that:

1) "Real" or ref counted copies are not that necessary, then

2) "Real" copies would still be good to have if, in the future, the plans would need to have mutable state or buffers (during execution).
(This could help the library to have less global state, but I understand if this is not a priority)

3) Reference counting still seems to me to be out of place here.

4) More importantly, from a C++ perspective (if you are interested at all), I can now finally declare `execute` as a `const` method with this knowledge.

5) In effect, because the of all this new knowledge (to me) I can safely implement this same thing with shared_ptr without waiting for this new `plan_copy` function to appear in my FFTW distribution. 

Thank you for the rapid answer.

matteo-frigo:
> I wonder if there are buffers behind the scenes in some global state of the library. And if so, how they are synchronized?

There are no buffers in plans.  When a plan needs a buffer it calls malloc(), or alloca() when the buffer is really small.

I think I agree with all your points 1-5 (see my Mar 2 comment), but other people feel otherwise.

correaa:
> > I wonder if there are buffers behind the scenes in some global state of the library. And if so, how they are synchronized?
> 
> There are no buffers in plans. When a plan needs a buffer it calls malloc(), or alloca() when the buffer is really small.

I am surprised about that also. (I am surprised often).

Unless I am missing something, once the library decides what plan to use, the amount of memory needed for execution seems to be determined. 
That is why I also assumed (wrongly again) that 1) allocation of the buffers happened on plan creation and 2) the plan would own the buffers.

> I think I agree with all your points 1-5 (see my Mar 2 comment), but other people feel otherwise.

Yes, this is IMO why it is so hard to write a good C++ wrapper for FFTW, because there are very subtle semantic implications that are hard to capture with simple classes.

The other problem with naive reference counting is that it is not thread-safe itself.
In `std::shared_ptr`, the counting is thread-safe; that is what makes it useful (beyond simply simulating garbage collection)

Anyway, food for thought for FFTW4 :) 

Cheers!

correaa:
> fftw plans are already "thread safe" in the sense that the plan is read-only when executing the plan, and thus you can execute the same plan from multiple threads. Plans do not store buffers, but they do store immutable arrays of trigonometric constants and other read-only information. So I don't think that the problem you mention needs to be solved.

sorry to insist, to clarify,

does plan creation allocate memory?
and, can plans be created from different threads simultaneously?

Issue #193, opened 2020-03-21
Test for -Wl,-no_compact_unwind not working


MarcusCalhoun-Lopez:
On older macOS systems, the [test](https://github.com/FFTW/fftw3/blob/c69d6c3c7512a814efe229d15e51dda0c1590ee6/configure.ac#L391) for `-Wl,-no_compact_unwind` does not actually test anything.
Even when the linker flag `-no_compact_unwind` is unsupported, the test succeeds because no linking is done (however, there is the warning `-no_compact_unwind: linker input file unused because linking not done`).

stevengj:
Closed by  #326

PR #326, opened 2023-05-14
Update autoconf m4 components

sevan:
Switch from AX_CHECK_COMPILER_FLAGS to AX_CHECK_COMPILE_FLAG & AX_CHECK_LINK_FLAG from autoconf-archive to resolve build issues with legacy Darwin.
Syntax has been updated to autoconf 2.69 format, this paves the way for updating to 2.71, but that needs to be handled separately as the remaining files in m4 need updating to current autoconf-archive versions, as well as changes to `configure.ac`.

Resolves issue #193


sevan:
@stevengj is there any issues with this proposed change?

stevengj:
LGTM

sevan:
Thanks :)

Issue #195, opened 2020-04-03
fftw3 export targets ${fftw3_lib}_omp failed


LilyWangL:
Hi,
I installed fftw3[openmp] successfully on Linux by VCPKG, and cmake created `libfftw3_omp.a libfftw3f_omp.a`. I found cmake create the cmake files as below:
![image](https://user-images.githubusercontent.com/47812810/78344588-2ead2280-75cf-11ea-9081-8358d0272453.png)
There no openmp cmake files, I add these codes to CMakeLists.txt and rebuild fftw3[openmp]. 
```
if (OPENMP_FOUND)		 
	export (TARGETS ${fftw3_lib}_omp NAMESPACE FFTW3:: FILE ${PROJECT_BINARY_DIR}/${fftw3_lib}_ompTargets.cmake)
	install(EXPORT ${fftw3_lib}_ompTargets
			NAMESPACE FFTW3::
			DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake
			COMPONENT Development)
endif()
```
fftw3[openmp] build failed with this error:
```
CMake Error: INSTALL(EXPORT) given unknown export "fftw3f_ompTargets"
CMake Warning:
  Manually-specified variables were not used by the project:

    VCPKG_PLATFORM_TOOLSET
    VCPKG_SET_CHARSET_FLAG
```
Could fftw3 generate target and config files for openmp? Thanks for any suggestion or help.

PR #331, opened 2023-08-06
Export all subtargets

dg0yt:
Export CMake config also for thread and omp targets when available.
For https://github.com/microsoft/vcpkg/issues/10194.
Resolves https://github.com/FFTW/fftw3/issues/195.


dg0yt:
Gentle ping.

PR #348, opened 2024-03-05
adding riscv cycle counter

ct-clmsn:
adds riscv cycle counter support to fftw3

### DIFF
--- a/kernel/cycle.h
+++ b/kernel/cycle.h
@@ -562,3 +562,25 @@ static inline ticks getticks(void)
 INLINE_ELAPSED(inline)
 #define HAVE_TICK_COUNTER
 #endif
+
+/*----------------------------------------------------------------*/
+/*
+ * RISC_V 64-bit cycle counter (RV64G)
+ */
+#if defined(__riscv) && (__riscv_xlen == 64) && !defined(HAVE_TICK_COUNTER)
+typedef unsigned long ticks;
+
+static __inline__ ticks getticks(void)
+{
+     ticks ret;
+
+     __asm__ __volatile__ ("csrrs %0, 0xc00, x0" : "=r" (ret) );
+
+     /* no input, nothing else clobbered */
+     return ret;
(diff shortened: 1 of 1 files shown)


rdolbeau:
The inclusion of the cycle counter (including support for rv32 in addition to rv64) is part of the various RISC-V "V" efforts.

ct-clmsn:
understood. this is a small fix and allows fftw3 to work on riscv processors that currently do not have vector hardware support. as of right now, the different riscv vector software support implementations are a block for using fftw3 on riscv.

kernel/cycle.h
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2003, 2007-14 Matteo Frigo
  * Copyright (c) 2003, 2007-14 Massachusetts Institute of Technology
+ * Copyright (c) 2024 Christopher Taylor, Tactical Computing Labs, LLC
```
stevengj, line 4: MIT will require a copyright assignment to merge PRs like this.

ct-clmsn: not an issue, can you send or post information on how that process works?

rdolbeau:
I understand the urgency; but I don't think this implementation is satisfactory for upstreaming:
(a) no support for RV32
(b) hard-to-understand csrr when there is dedicated aliases like rdcycle and rdtime
(c) if this is using rdcycle, then user-land access was deprecated (security...) and will no longer be working Real Soon Now. Recent kernels on recent hardware have already dropped support (e.g. the manufacturer's kernel on the SpacemiT K1 SoC at least).

If there's some urgency, I would suggest merging separately the RISC-V version from [here](https://github.com/rdolbeau/fftw3/blob/84fd9d30380129f8abdd360e3329b905dd824613/kernel/cycle.h#L566-L589).

rdolbeau:
#361 was merged, so I think we can drop this one

PR #349, opened 2024-03-18
Include missind header for Win32 Threads check

BtbN:
Recent compilers (mostly clang) just abort on implicit function declaration. So this check could never succeed there, since the proper header for the tested function was not included.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -700,7 +700,8 @@ if test "$enable_threads" = "yes"; then
         # Win32 threads are the default on Windows:
     if test -z "$THREADLIBS"; then
         AC_MSG_CHECKING([for Win32 threads])
-        AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <windows.h>]], [[_beginthreadex(0,0,0,0,0,0);]])],[THREADLIBS=" "; AC_MSG_RESULT(yes)],[AC_MSG_RESULT(no)])
+        AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <windows.h>
+                                          #include <process.h>]], [[_beginthreadex(0,0,0,0,0,0);]])],[THREADLIBS=" "; AC_MSG_RESULT(yes)],[AC_MSG_RESULT(no)])
     fi
 
     # POSIX threads, the default choice everywhere else:


configure.ac
```diff
@@ -700,7 +700,7 @@ if test "$enable_threads" = "yes"; then
         # Win32 threads are the default on Windows:
     if test -z "$THREADLIBS"; then
         AC_MSG_CHECKING([for Win32 threads])
-        AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <windows.h>]], [[_beginthreadex(0,0,0,0,0,0);]])],[THREADLIBS=" "; AC_MSG_RESULT(yes)],[AC_MSG_RESULT(no)])
+        AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <windows.h>]], [[#include <process.h>]], [[_beginthreadex(0,0,0,0,0,0);]])],[THREADLIBS=" "; AC_MSG_RESULT(yes)],[AC_MSG_RESULT(no)])
```
stevengj, line 703: both headers should be part of the same "prologue" argument to [`AC_LANG_PROGRAM`](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Generating-Sources.html):
```suggestion
        AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <windows.h>
#include <process.h>]], [[_beginthreadex(0,0,0,0,0,0);]])],[THREADLIBS=" "; AC_MSG_RESULT(yes)],[AC_MSG_RESULT(no)])
```

stevengj: Or maybe we should just replace `windows.h` with `process.h` if the latter is the correct header. 
```suggestion
        AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <process.h>]], [[_beginthreadex(0,0,0,0,0,0);]])],[THREADLIBS=" "; AC_MSG_RESULT(yes)],[AC_MSG_RESULT(no)])
```
(I thought that `windows.h` used to include `process.h`?  But the [`_beginthreadex` docs](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/beginthread-beginthreadex?view=msvc-170) say that `process.h` is the right header for this function.)

BtbN: I'm not sure if it's valid to include process.h without windows.h.
Typically other winapi headers depend on windows.h having already been included.

Though I tested it, and plain replacing windows.h with process.h here does _work_ at least.

stevengj: In [thread.c](https://github.com/FFTW/fftw3/blob/master/threads/threads.c#L164-L165) we use both headers, so maybe it would be safer to do so here, i.e. using the first version of my patch above.

BtbN: Yeah, that's what I went with as well.

PR #354, opened 2024-05-06
cmake: Fix compilation with VisualStudio clang-cl

antoniovazquezblanco:
While contributing a wrap file to use fftw3 in meson (https://github.com/mesonbuild/wrapdb/pull/1503) I found that current CMakeLists.txt fails to build with VisualStudio clang-cl. This fixes the issue.


stevengj:
LGTM

PR #363, opened 2024-08-22
Add SIMD optimization for LoongArch

Pengzhou0810:
These four patches have added SMID optimization for the loongarch（a new architecture of Loongson CPU）

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -99,6 +99,11 @@ LSX_LIBS = dft/simd/lsx/libdft_lsx_codelets.la	\
 rdft/simd/lsx/librdft_lsx_codelets.la
 endif
 
+if HAVE_LASX
+LASX_LIBS = dft/simd/lasx/libdft_lasx_codelets.la	\
+rdft/simd/lasx/librdft_lasx_codelets.la
+endif
+
 if HAVE_GENERIC_SIMD128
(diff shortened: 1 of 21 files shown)


Pengzhou0810:
Hello @stevengj , could you please help review my patch and see if it can be merged into fftw3.

rdolbeau:
Hello,
couple of quick comments:
(a) split the rdticks stuff in its own PR? as it's much smaller and easier, and has value for all users of the architecture (w/ or w/o SIMD)
(b) there seem to be different SIMD implementations, what is the difference between them? Are they both optional or mandatory in the architecture, what are their respective supported width and datatypes, etc, so it's easier to understand the PR.
Cordially,

yinshiyou:
> Hello, couple of quick comments: (a) split the rdticks stuff in its own PR? as it's much smaller and easier, and has value for all users of the architecture (w/ or w/o SIMD) (b) there seem to be different SIMD implementations, what is the difference between them? Are they both optional or mandatory in the architecture, what are their respective supported width and datatypes, etc, so it's easier to understand the PR. Cordially,

Thank you very much for your reply. To resolve the conflicts , I update a new PR  #376 with my account. This PR will be closed.
a) I am not sure what you mean by "rdticks stuff", could you please comment under the new PR #376 .
b) In the new PR, I add a few notes for this new SIMD instruction.

Issue #371, opened 2024-10-08
RVV/R5V implementation and optimization


ken-unger:
Hi @rdolbeau 

We’ve recently been using your risc-v optimized fftw3 implementation found in https://github.com/rdolbeau/fftw3/tree/riscv-v-clean

I’m interested in your assessment of the opportunity for further optimization of the vector implementation within simd-support/simd-r5v.h.   Is there additional work that might make sense here, or do you believe you have achieved most of the gains?  For example, I see that lmul=1 (m1) is used whereas I would have anticipated a higher lmul might be possible.  Or perhaps that has been explored already and rejected.

Any thoughts appreciated.

Thanks,
Ken

rdolbeau:
FFTs (including FFTW3's codelets) tend to require a lot of registers unless quite small. Anything that increase register pressure, including split-mode and larger LMUL (which is effectively equivalent to unrolling for this use case), are likely to cause excessive spillage of registers which will degrade performance. Also, some instructions do not like larger LMUL performance-wise - `vrgather` in particular is likely to misbehave in many implementations (e.g. it takes a time quadratic with LMUL on the SpacemiT K1).

Nonetheless, there's certainly room for improvements (it has only been tested on my K1), but I'd say currently the compilers should be the primary target. I'm attaching a subset of benchmarks on the K1, and as a remark:

(a) gcc (blue/cyan) is better than LLVM (red/orange) overall
(b) gcc is better when using the 'interleaved' representation ('r5v') than when using the 'split' representation ('r5vsplit')
(c) llvm is better when using the 'split' representation than when using the 'interleaved' one

There should be a more consistent behavior of the software stack before doing much more tuning on the library itself. This is gcc-14 and clang-16 from a recent Debian Sid, so it's not an up-to-date issue (though LLVM moves quickly). A quick glance at the ASM suggest that excessive spillage is already an issue in some of the slower variants. Under a similar test setup, the SVE implementation is much more stable in behavior across compilers.

[doit.ALL.double.1d.drxx.p2.pdf](https://github.com/user-attachments/files/17291023/doit.ALL.double.1d.drxx.p2.pdf)

matteo-frigo:
Register pressure is a complicated issue for which we had a more-or-less "optimal" solution that worked at some point and may have been broken by more recent compilers.

[...]

PR #377, opened 2025-02-05
Merging of RISC-V V1.0 support.

rdolbeau:
This is a cleaned-up version of branch riscv-v from rdolbeau/fftw3.

No scalability, this is using multiple sets of masked operations to support all width multiple of 128 bits.
This goes up to 16384 bits to support the EPAC VPU design.

This is now using the official V1.0 intrinsics, and was tested with gcc-14 and clang-18. It can bes tested on qemu, or on hardware with V1.0 support. It was thoroughly tested on a BananaPi F3 with a Spacemit K1 SoC, which supports V1.0 with 256-bits registers.

The code was originally based on the SVE code and used EPI intrinsics. It was further refined on my bannapi f3, and some improvements were inspired by the code from #279, and is faster overall on the F3 (see a sample graph in #371)

It's currently still using the generated vtw.h, it still need to use the same macro-based approach as upstream's merged SVE (see commit 126e3b9)


rdolbeau:
@matteo-frigo :

This contains both the "standard" implementation (with real & imaginary part stored interleaved in a single vector register) and the "split" implementation (with the real and imaginary part stored in two different registers). However, only the "standard" is sued for now - testing the split requires the manual step of copying over the file. In theory, both could be enabled at the same time, but that would be a lot of codelets. The "split" is very register-intensive doesn't play well with larger codelets, hence the choice of "standard" by default. Feedback welcome.

rdolbeau:
Test release (with all codelets pre-generated) available in https://github.com/rdolbeau/fftw3/releases/tag/r5v-test-release-005

sh-zheng:
Great work!

This branch is now using "r5v" as its arch tag. May the "rvv" be relatively more often used? Or according to the "V" specificion, the Zvl128b/256b... may be more official?

I find that the arch tag in OpenBLAS uses Zvl128b/256b

simd-support/simd-r5v.h
```diff
@@ -0,0 +1,473 @@
...
+}
+
+static inline V FLIP_RI(const V x) {
+	Vint idx = TYPEINT(vid_v)(2*VL); // (0, 1, 2, 3, ...)
+	//Vint vone = TYPEINT(vmv_v_x)(1, 2*VL);
+	Vint hidx = TYPEINT(vxor_vx)(idx, 1, 2*VL); // (1, 0, 3, 2, ...)
+	return TYPE(vrgather_vv)(x, hidx, 2*VL);
+}
+
+#define VNEG(a)   TYPE(vfsgnjn_vv)(a,a,2*VL)
```
sh-zheng, line 136: Could be simplified as vfneg_v

simd-support/simd-r5v.h
```diff
@@ -0,0 +1,473 @@
...
+//#define VONE  VLIT1(DS(1.,1.f))
+//#define VMINUSONE VLIT1(DS(-1.,-1.f))
+// generate (all 1, 0, all 1, 0, ...) to split real and imagine parts
+#define VPARTSPLIT TYPEINT(vsub_vx)(TYPEINT(vand_vx)(TYPEINT(vid_v)(2*VL), 1, 2*VL), 1, 2*VL)
+
+static inline V VDUPL(const V x) {
+	Vint idx = TYPEINT(vid_v)(2*VL); // (0, 1, 2, 3, ...)
+	//Vint vnotone = TYPEINT(vmv_v_x)(DS(~1ull,~1), 2*VL);
+	Vint hidx = TYPEINT(vand_vx)(idx, DS(~1ull,~1), 2*VL); // (0, 0, 2, 2, ...)
+	return TYPE(vrgather_vv)(x, hidx, 2*VL);
```
sh-zheng, line 120: The index of destination of vrgather_vv depends on the value in vector reg, which makes it relatively unpredictable for compiler/microarchitecture. May the vslide instruction have a more regular behaviour and better performance?

simd-support/simd-r5v.h
```diff
@@ -0,0 +1,473 @@
...
+
+static inline V VZMULIJ(V tx, V sr)
+{
+    V tr = VDUPL(tx);
+    V ti = VDUPH(tx);
+    ti = VMUL(ti, sr);
+    sr = VBYI(sr);
+    return VFMA(tr, sr, ti);
+}
+#else
```
sh-zheng, line 212: VZMULJ, VZMULI, and VZMULIJ in else macro are the same as them in if macro, so they may be coalesced

simd-support/simd-r5v.h
```diff
@@ -0,0 +1,473 @@
...
+#define VFMSCONJ(b,c)  VFMACONJ(b,VNEG(c)) // fixme: improve
+#define VFNMSCONJ(b,c) VNEG(VFMSCONJ(b,c)) // fixme: improve
+
+#if 1
+static inline V VCONJ(const V x) {
+	Vint idx = TYPEINT(vid_v)(2*VL); // (0, 1, 2, 3, ...)
+	//Vint vone = TYPEINT(vmv_v_x)(1, 2*VL);
+	Vint hidx = TYPEINT(vand_vx)(idx, 1, 2*VL); // (0, 1, 0, 1, 0, 1)
+	//return TYPEMASK(vfsgnjn_vv)(x, x, x, INT2MASK(hidx), 2*VL);
+	return TYPEMERGEDMASK(vfsgnjn_vv)(INT2MASK(hidx), x, x, x, 2*VL);
```
sh-zheng, line 155: The mask register is scarce resource in rvv, and mask undisturbed policy may hinder out-of-order execution. So the unmask version may have benefits.

matteo-frigo:
Can we take a more "practical" approach to the vector length, and only support lengths for which silicon actually exists, and maybe twice that for good measure?  Or is there any actual 16384-bit silicon?

If you could replace the vtw.h thing with the new macros it would be great.

rdolbeau:
> Can we take a more "practical" approach to the vector length, and only support lengths for which silicon actually exists,
> and maybe twice that for good measure? Or is there any actual 16384-bit silicon?

There is, actually, though it's research silicon for now. The European Processor Initiative (EPI) test chips EPAC v1.0 and [v1.5](https://www.european-processor-initiative.eu/european-processor-initiative-announces-the-successful-bring-up-of-the-epac1-5-acceleration-chip/) contain a 16384-bits (VL=256 @ 64 bits) vector unit (VPU). It's only v0.7.1 of the specifications in those, but the next tape-out in a different European project will support v1.0 instead. In fact, it's the hardware for which the code was originally written (way back with EPI-specifics intrinsics). There's also a FPGA-based Software Development Platform for EPI partners.

That being said, I agree that he wider vectors are not likely to be useful to the general public and are bloating the libraries. A solution could be to add a "--max-scalable-width" option for SVE and RISC-V, probably defaulting to 512 (A64FX width), that would only build the codelets up to that size. That would avoid the bloat of unused codelets. The extra files/directories needed to support the additional width in the source code are quite small and very similar so probably not an issue. Then users with specifics requirements can rebuild with the appropriate value for more "exotic" hardware.

> If you could replace the vtw.h thing with the new macros it would be great.

I'll try to do it ASAP, but I wanted to bring the code forward now that SVE is merged (so the principle seems sound) and hardware is actually available commercially, even if with limited width.

I do still wonder if the scalable vector, or in fact all SIMD with proper mask support, couldn't be used more efficiently. Given the current usage model of SIMD in FFTW3, in theory it should be possible to implement the codelets with the internal loop in scalable/masked mode, i.e. replace (here from n1fv_16.c):
```
/* VL is a compile-time constant */
for (i = v; i > 0; i = i - VL, xi = xi + (VL * ivs), xo = xo + (VL * o
vs), MAKE_VOLATILE_STRIDE(32, is), MAKE_VOLATILE_STRIDE(32, os)) {
/* code goes here */
```
by something like
```
/* dynamic_VL set to the hardware length */
for (i = v; i > 0; i = i - dynamic_vl, xi = xi + (dynamic_VL * ivs), xo = xo + (dynamic_VL * o
vs), MAKE_VOLATILE_STRIDE(32, is), MAKE_VOLATILE_STRIDE(32, os)) {
/* here, build a mask so if (v % dynamic_VL) is non-zero the last iteration is appropriately masked */
/* masked code goes here */
```

Alternatively, it could be two loops, an unmasked main loop while `i` is larger or equal than `dynamic_VL`, and a single-iteration, masked reminder loop.

However, VL is baked into other places used by the planner infrastructure, and I'm not sure how easy/achievable that would be. But it would be cleaner and would enable more use cases for SIMD, as this can be used to implement any number of iterations rather than just multiple of VL. AVX-512 (and newer) would also potentially benefits has it can be masked just as well.

matteo-frigo:
Re: 16384 bits.   I would do something like --enable-sve enables <= 512 or whatever is a reasonable upper bound today, and --enable-large-sve enables <= 16384.  Basically pretend it's another SIMD architecture.  I don't feel strongly about this, though.  Let's try to capture the practical common case + the exceptional case.

Re: dynamic VL.  As you may have surmised, our SIMD theory is a bit long in the tooth.  We hacked something when there was only SSE and 3DNow!, then SSE2 came and the rest is history.  I would probably leave the existing SIMD stuff alone and create a new set of codelets.  Thus we could have scalar/, simd/, and dynsimd/ codelets.  The scalar/ codelets are already "vectorized" in the sense that they compute a vector of transforms, not just one.  If I am reading what you say correctly, the dynsimd/ codelets would be the same as scalar, except perhaps with some restrictions on the strides.  Every "genus" of codelets defines a function okp() that tells the planner whether or not the codelet is applicable, so this restriction would be easy to capture.

simd-support/simd-r5v.h
ChipKerchner: vfneg.v vd, v[op1] // == vfsgnjn.vv vd, v[op1], v[op1]

rdolbeau:
@matteo-frigo Sorry, real life interfered and I've forgotten to deal with this. Now, someone (@hugomeiland) has proposed a patch to my branch which implements you request, and after looking at it, it LGTM (it's mostly mechanical stuff and alignment with the current SVE version) and I'm inclined to merge it on my branch. However, the patch being written with the help of AI, I'm not sure that it will be acceptable to FFTW in general.

PR #382, opened 2025-04-24
Use modern `cmake` syntax for `OpenMP`

### LINKED ISSUE #381: Compile failure with Cmake 4.0 due cmake_minimum_required <3.5 support removed
Due to [this](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features) upstream change, fftw3 will not compile with cmake 4.0 - with the error:

```CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
Error:   Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.```

See also [this](https://fedoraproject.org/wiki/Changes/CMake4.0) note from fedora linux about the cmake change. Not sure if any linux distros use cmake instead of autoconf to build, though it is at least used by vcpkg which is where I encountered it.

### PR DESCRIPTION
**Pull request by ywwry66**
Cf. https://cliutils.gitlab.io/modern-cmake/chapters/packages/OpenMP.html

Also fixes #381.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.0)
+cmake_minimum_required (VERSION 3.9)
 
 if (NOT DEFINED CMAKE_BUILD_TYPE)
   set (CMAKE_BUILD_TYPE Release CACHE STRING "Build type")
@@ -123,7 +123,7 @@ if (Threads_FOUND)
 endif ()
 
 if (ENABLE_OPENMP)
-  find_package (OpenMP)
+  find_package (OpenMP COMPONENTS C REQUIRED)
 endif ()
 if (OPENMP_FOUND)
   set (HAVE_OPENMP TRUE)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by arrowd — APPROVED** [review]
As a heavy CMake user I can say that the change makes sense.

Note that the contemporary FindOpenMP module documentation says that the `COMPONENTS C` part was added in CMake 3.10, not 3.9: https://cmake.org/cmake/help/v4.3/module/FindOpenMP.html

File: CMakeLists.txt [hunk 1]
```diff
@@ -366,7 +366,7 @@ if (OPENMP_FOUND)
   target_link_libraries (${fftw3_lib}_omp ${fftw3_lib})
   target_link_libraries (${fftw3_lib}_omp ${CMAKE_THREAD_LIBS_INIT})
   list (APPEND subtargets ${fftw3_lib}_omp)
-  target_compile_options (${fftw3_lib}_omp PRIVATE ${OpenMP_C_FLAGS})
+  target_link_libraries (${fftw3_lib}_omp OpenMP::OpenMP_C)
```
stevengj, line 369: Aren't there usually compiler flags too, in addition to the link libraries?

stevengj: Oh nevermind, I see that [cmake import targets](https://cmake.org/cmake/help/v4.3/manual/cmake-buildsystem.7.html#imported-targets) nowadays also modify the compiler flags automatically.

PR #386, opened 2025-08-26
bump cmake min required version

iboB:
CMake support for versions <3.10 is deprecated and to be removed soon. Configuring fftw3 with the current min version leads to lengthy CMake warnings which clog the configure log.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.0)
+cmake_minimum_required (VERSION 3.10)
 
 if (NOT DEFINED CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


egpbos:
Alternative would be to change to a more explicit range of compatible versions: `cmake_minimum_required (VERSION 3.0...4.0)`. This would have the benefit of also fixing https://github.com/FFTW/fftw3/issues/381 which means one can again build with CMake 4 (which currently doesn't work).

jschueller:
+1

stevengj:
Seems like we might as well merge this since it is causing problems, and we don't care that much about supporting old `cmake` versions anyway.

stevengj:
(The primary build system for FFTW continues to be autoconf/configure … `cmake` was mainly included for Windows users.)

egpbos:
This fixes #381.

PR #391, opened 2025-09-25
Improve NEON support on MacOS

green-br:
Add arm host_cpu to NEON check found on MacOS.


fxcoudert:
Why `arm` and not `arm64` (which is what Apple `uname -m` outputs)?

green-br:
Seems to be what `config.guess` returns.  Your suggestion to use `--build=aarch64-apple-darwin#{OS.kernel_version}` in the home-brew formula as suggested`Homebrew/homebrew-core/pull/245725` means we can close this ticket.

PR #397, opened 2025-12-04
[WIP] Meson build system

mia-0:
Since CMake has too many problems to list and Autotools doesn’t really integrate with anything, I decided to port most of the Autotools build system to Meson. This includes maintainer mode. I did not touch any existing files other than for the commit that removes CMake, so this is not as clean and simple as it could be.

The genfft part deviates a bit from Autotools by using (requiring) [Dune](https://dune.build/), which will download and compile known-working versions of the OCaml toolchain and any required packages as necessary.

See `README-meson.md` in the top-level directory for more information.

#### Targets

GCC 15 with glibc except as noted. Cross-compiled builds were tested by running the `bench` program through a few cases using QEMU binfmt wrappers or WINE and making sure it also picked a few SIMD code paths.

* x86_64 native
* aarch64 cross (Android NDK r28c, Clang)
* loongarch64 cross
* ppc64le cross
* riscv64 cross
* s390x cross
* MinGW-w64 x86_64 cross

#### Environments

* plain
* KDevelop
* VSCode with Meson extension

#### Other

* Creating dist tarballs
* Including FFTW as subproject of another project via [Meson’s Wrap system](https://mesonbuild.com/Wrap-dependency-system-manual.html)

### On Windows 11:

MSVC cl 19.50 / link 14.50. Other than MSVS 2026, only Meson and Ninja were installed using the MSI package from [Meson’s Release page](https://github.com/mesonbuild/meson/releases). No WSL or other external dependencies.

Static library build only because we’re not explicitly exporting some internal symbols that are needed to dynamically link threads/omp/mpi libs and tools to the main library. But yes, it basically works with MSVC-style toolchains as well. Was easier than I thought, too.

#### Targets

* x86 native
* x86_64 native

#### Environments

* plain (developer console)
* Visual Studio 2026 (Meson vs2022 backend)
* Visual Studio 2026 (vslite project setup)

---

I also wanted to test some BSDs but didn’t get around to it yet. As for macOS and other Apple targets, Meson does have an Xcode backend, but I was unable to test it because I do not have a Mac and my aging travel laptop took more than half an hour just to boot a macOS VM (it really is not meant to run on anything but Apple hardware). Help is appreciated.

## To Do

These are required to reach feature parity with Autotools:

- [ ] Fortran wrappers (needs replacements for Autoconf-provided macros)
- [ ] ARM v7a/v8 performance counters
- [ ] Commercialized tarballs

Bonus features:

- [ ] Build multiple precisions at once
- [ ] Replace Perl wrapper with proper Meson test and benchmark suites
- [ ] Generate more complete set of CMake support files (please somebody else do this)


antoniovazquezblanco:
Awesome effort! Would love to see it merged! Thank you! :D

mia-0:
Actually I don’t care anymore. FFTW3 is not free software (centers people), it is a product (centers market value). I have no interest in supporting the latter, and suggest taking a look at FFmpeg’s FFT implementations instead of using FFTW3.

PR #405, opened 2026-05-26
Elbrus2000 architecture support

arrowd:
`make check` passes.

### DIFF
--- a/kernel/cycle.h
+++ b/kernel/cycle.h
@@ -524,6 +524,18 @@ INLINE_ELAPSED(inline)
 #define HAVE_TICK_COUNTER
 #endif
 
+#if defined(__e2k__) && !defined(HAVE_TICK_COUNTER)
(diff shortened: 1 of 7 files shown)


matteo-frigo:
Thanks.

I am confused by your PR.  I didn't know about e2k (or perhaps I saw an announcement at the time and I forgot), but as far as I can tell it is a SPARC derivative.  How can it have SSE2/AVX?

arrowd:
> as far as I can tell it is a SPARC derivative. 

No it is an "old" Elbrus that was a SPARC derivative. Elbrus2000 (e2k) is a completely original architecture, a VLIW one.

> How can it have SSE2/AVX?

The compiler ships SSE/AVX compatible headers (*intrin.h) which export intrinsics that are expanded to e2k SIMD instructions. This way e2k can piggy-back on the x86 SIMD code in various projects.

PR #410, opened 2026-07-01
added pkg-config support for MPI builds

### LINKED ISSUE #57: RFE: fftw.pc.in conditional MPI library inclusion
I understand that the position of those working on FFTW is to minimize library dependencies (at least with regards to `pthreads` and `fftw3*_threads`), but when building FFTW with MPI enabled, it may be desirable to include `-lfftw3*_mpi` in the pkg-config libraries portion. I'm not sure if this may cause problems on other systems, but at least in Fedora, the `fftw3*.pc` files will be installed in their respective MPI `PKG_CONFIG_PATH` (i.e. `/usr/lib64/mpich/lib/pkgconfig` or similar) and will not conflict with those non-MPI pc files when the user loads the corresponding MPI module (i.e. `module load mpi/mpich-x86_64`) for configuring and building with MPI.

In any case, an easy way to conditionally include `-lfftw3*_mpi` is in the `configure.ac` file add under the `if test "$enable_mpi" = "yes"; then` test the lines

`LIBFFTW3MPI=-lfftw3${PREC_SUFFIX}_mpi`
`AC_SUBST(LIBFFTW3MPI)`

and change the `Libs:` in `fftw.pc.in` to
`Libs: -L${libdir} @LIBFFTW3MPI@ -lfftw3@PREC_SUFFIX@ @LIBQUADMATH@`.

MarDiehl:
will create additional pkgconfig file, following the design of FFTW.

closes #57 if I understand the issue correctly.

Note that this change probably is in conflict with #407


MarDiehl:
#407 uses separate `.pc.in` files, this uses sed to adjust the existing file. From my perspective, it would make sense to use the same approach for MPI, OpenMP, and threads.

